PR #425, opened 2014-04-22
Expose RiakConnectionPool available and inUse

kristomasette:
I added a couple of getter messages so that I can monitor.


alexmoore:
Hi, the 1.X series Riak Java Client is no longer supported. We appreciate your PR but cannot accept it. 

Thanks,
Alex

Issue #481, opened 2014-10-14
Refine how you get data from the FetchValue.Response [JIRA: CLIENTS-35]


Basho-JIRA:
There are a couple things that can be improved in the FetchValue.Response in terms of how you get the returned data from it. 

* Convenience method to return RiakObject(s) 
* Additional methods to allow passing in a ConflictResolver and/or a Converter rather than registering them with their respective factories.

PR #498, opened 2015-02-25
Gh481 response options

broach:
You can now explicitly supply a `Converter<T>` and `ConflictResolver<T>`
rather than being required to register them with the associated
factories.

Addresses #481 (CLIENTS-35)


zeeshanlakhani:
:+1: e7caf41. Only the MR test, already discussed, is failing (and it's already failing on develop). TIL about lowercased index names.

PR #506, opened 2015-03-03
Testing bucket types [JIRA: CLIENTS-216]

christophermancini:
I added a prefix to the bucket types used with the client libs tests to prevent collisions / conflicts with other client libs when developing / testing in multi-client environments.


broach:
create jira issue from this pr

christophermancini:
LOL, can't believe I did that. This is why PRs exist. :)

I will fix it Monday.

broach:
:+1:

PR #509, opened 2015-03-23
Declare constants as final - use double-checked locking for lazy init

nowheresly:
For performance consideration,it is possible to remove the synchronized keyword for 3 methods inside RiakObject class without compromising thread safety.
See for reference: http://stackoverflow.com/questions/17164454/why-double-checked-locking-is-25-faster-in-joshua-bloch-effective-java-example?rq=1

### DIFF
--- a/src/main/java/com/basho/riak/client/core/query/RiakObject.java
+++ b/src/main/java/com/basho/riak/client/core/query/RiakObject.java
@@ -298,14 +298,22 @@ public final class RiakObject
      * @return the {@link RiakIndexes} that encapsulates any/all indexes for this RiakObject
      * @see <a href="http://docs.basho.com/riak/latest/dev/advanced/2i/">Riak Secondary Indexes</a>
      */
-    public synchronized RiakIndexes getIndexes()
+    public RiakIndexes getIndexes()
     {
         // Lazy initialization of the internal container.
-        if (null == riakIndexes)
+        RiakIndexes result = this.riakIndexes;
+        if (null == result)
         {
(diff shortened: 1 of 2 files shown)


alexmoore:
While Joshua Bloch's book is great, it's getting a little outdated.  Java Concurrency in Practice lists double checked locking as an antipattern (see Page 349, Listing 16.7. "Double-checked-locking antipattern. Don’t do this.") 

The way it's done is the recommended simple thread-safe way.  

Thanks for the PR though!

nowheresly:
As stated in Java Concurrency in Practice, page 349 just above listing 16.7:

> Subsequent changes in the JMM (Java 5.0 and later) have enabled DCL to work if resource is made **volatile**, and the performance impact of this is small since volatile reads are usually only slightly more expensive than nonvolatile reads.

Since the fields are declared `volatile`, my PR is correct as far as threading issue is concerned and thus cannot be compared to the listing 16.7 where the field is **not** declared as `volatile`.

In the same chapter, paragraph 16.2.3, Brian Goetz explains when it is correct to use a `synchronized`method to achieve lazy loading:

> UnsafeLazyInitialization can be fixed by making the getResource method synchronized, as shown in Listing 16.4. Because the code path through getInstance is fairly short (a test and a predicted branch), **if getInstance is not called frequently by many threads**, there is little enough contention for the SafeLazyInitialization lock that this approach offers adequate performance.

It's ok for me to consider `getIndexes`, `getUserMeta`and `getLinks`as being **fairly short and not called frequently by many threads** so we may close this PR.

Thanks for taking time review it.

Issue #513, opened 2015-04-11
mvn install fails on Java8 [JIRA: CLIENTS-325]


srgg:
When I tried to build the latest java client with java8 (With java7 everything works well.), I've got the following error:

```
[ERROR] Failed to execute goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.9:check (default) on project riak-client: Execution default of goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.9:check failed. IllegalArgumentException -> [Help 1]
```

code was successfully compiled, but signature check has failed.

 Most likely this error is caused by the incomtability of animal-sniffer plugin with java8

Here is java version i've used:

```
 java -version
java version "1.8.0_25"
Java(TM) SE Runtime Environment (build 1.8.0_25-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.25-b02, mixed mode)
```

PR #514, opened 2015-04-11
CLIENTS-325 animal-sniffer-maven-plugin version was upraded to support j...

srgg:
...ava8 builds


srgg:
It resolves #513 (CLIENTS-325)

broach:
:+1:

PR #516, opened 2015-04-14
DP-46  GetCoverage operation to obtain either full or key coverage

srgg:
This PR utilizes API introduced by DP-6. Therefore **IT might BE MERGED ONLY when** DP-6 will be completed and all the following PRs will be merged:
-  https://github.com/basho/riak_pb/pull/106
-  https://github.com/basho/riak_api/pull/84
-  https://github.com/basho/riak_kv/pull/1087

**NOTE:**  As soon as riak_pb version with the PR mentioned above to be released, **pom.xml should be updated to use  released version of riak_pb**. Currently, since there is no released version with the required functionality, SNAPSHOT dependecy is used.

### DIFF
--- a/src/main/java/com/basho/riak/client/api/commands/kv/GetCoverage.java
+++ b/src/main/java/com/basho/riak/client/api/commands/kv/GetCoverage.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2013 Basho Technologies Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.basho.riak.client.api.commands.kv;
+
+import com.basho.riak.client.api.RiakCommand;
+import com.basho.riak.client.api.commands.CoreFutureAdapter;
+import com.basho.riak.client.core.RiakCluster;
+import com.basho.riak.client.core.RiakFuture;
+import com.basho.riak.client.core.operations.GetCoverageOperation;
+import com.basho.riak.client.core.query.Location;
+import com.basho.riak.client.core.query.Namespace;
+import com.basho.riak.client.core.util.BinaryValue;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Command used to get a coverage information from Riak.
+ * <p>
+ * It returns coverage for the specified {@link com.basho.riak.client.core.query.Location} or
+ * {@link com.basho.riak.client.core.query.Namespace}.
+ * if nothing was specified then the full coverage will be returned.
+ * <p>
+ * Coverage is returned as a {@link List} of Riak entry points (Riak nodes), which are defined as a {@link java.util.Map.Entry}:
+ * <ul>
+ *  <li>{@link Map.Entry#getKey()} contains address of the corresponding endpoint</li>
+ *  <li> {@link Map.Entry#getValue()} contains port</li>
+ * </ul>
+ *
+ * <b>Note:</b> The endpoint's address is always returned as IPv4 in dot-decimal notation.
+ *
+ * @author Sergey Galkin <sgalkin at basho dot com>
+ */
+public class GetCoverage extends RiakCommand<GetCoverage.Response, String>{
+    private final Namespace namespace;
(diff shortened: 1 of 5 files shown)


srgg:
@broach  could you please perform a kind of pre-review of my PR, just to be sure that everything was implemented in a proper way

broach:
The code for the operation aside, if I were to be implementing this I would probably make a "(sorta) smart routing" NodeManager implementation with a caching mechanism of some sort. Since the client design is based on the NodeManager deciding which node to execute command on, this seems like a logical fit. 

Exposing this as a user facing command doesn't make a lot of sense to me because of that, unless you want to then also make it so they can supply a hint with a given Fetch/Store ... which just seems like making the user do stuff they shouldn't have to worry about because it could be done how I describe above.

srgg:
I will take a look how such NodeManager might be implemented. Thanks for pointing it out to me

seancribbs:
Also, I haven't fully reviewed the server-side bits, which I promised to do after the 2.1 release.

src/main/java/com/basho/riak/client/api/commands/kv/GetCoverage.java
```diff
@@ -0,0 +1,182 @@
...
+ * </ul>
+ *
+ * <b>Note:</b> The endpoint's address is always returned as IPv4 in dot-decimal notation.
+ *
+ * @author Sergey Galkin <sgalkin at basho dot com>
+ */
+public class GetCoverage extends RiakCommand<GetCoverage.Response, String>{
+    private final Namespace namespace;
+    private final BinaryValue key;
+    private final boolean forceUpdte;
```
hmmr: 'a' is missing in "forceUpdate" (globally).

src/main/java/com/basho/riak/client/core/operations/GetCoverageOperation.java
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright 2013 Basho Technologies Inc
```
hmmr: Update the year to 2015?

src/test/java/com/basho/riak/client/api/commands/itest/ITestGetCoverage.java
```diff
@@ -0,0 +1,106 @@
...
+        /**
+         * Since GetCoverage command is the only way to get an information about the number of available Riak nodes
+         * there is no chance to verify count of returned EPs. Therefore we will iterate through the list of the
+         * returned EPs and ping each of them to ensure that at least a valid EP was returned.
+         * */
+        verifyThatAllNodesAreAvailable(response.entryPoints());
+    }
+
+    private static void verifyThatAllNodesAreAvailable(Collection<Map.Entry<String,Integer>> nodes) throws UnknownHostException, InterruptedException {
+        for(Map.Entry<String,Integer> e: nodes){
```
hmmr, line 83: As the actual call can return entry points with "not_routed" instead of a real IP address, be sure to not send any pings to it.

src/main/java/com/basho/riak/client/core/operations/GetCoverageOperation.java
```diff
@@ -0,0 +1,139 @@
...
+        private final List<Map.Entry<String,Integer>> entryPoints;
+
+        public Response(List<RiakKvPB.RpbApiEp> epList) {
+            entryPoints = new ArrayList<Map.Entry<String,Integer>>(epList.size());
+            for(RiakKvPB.RpbApiEp ep: epList){
+                entryPoints.add(new AbstractMap.SimpleEntry<String, Integer>(ep.getAddr().toStringUtf8(), ep.getPort()));
+            }
+        }
+
+        public List<Map.Entry<String,Integer>> geEntryPoints() {
```
hmmr: 't' missing in "getEntryPoints".

src/main/java/com/basho/riak/client/api/commands/kv/GetCoverage.java
```diff
@@ -0,0 +1,182 @@
+/*
+ * Copyright 2013 Basho Technologies Inc
```
hmmr: It's 2015 now.

src/test/java/com/basho/riak/client/api/commands/itest/ITestGetCoverage.java
srgg, line 83: @hmmr If I'm not mistaken we decided do not return an information about endpoints that can't be used.

hmmr, line 83: @srgg In fact, we 'respectfully disagreed' on the matter, but I have since become convinced that not_routed makes little sense without indication on which node the failing listener is. I have made changes to filter such non-addresses in the returned results, and updated my PRs in riak_api and riak_kv.

srgg, line 83: It seems that we have reached a consensus.

alexmoore:
Closed per conversation with Sergey, Matt, and Oleg.

PR #517, opened 2015-04-30
READY: Fix for Java8 builds, remove tools.jar dependency. 

alexmoore:
- Fixed a few tests that had issues
- Updated Jacoco so it will build on Java8
- Removed dependency-reduced-pom.xml, as it's generated by maven
- Ensure tools.jar isn't a hardcoded path.


christophermancini:
:+1: 9fe8a45

Issue #518, opened 2015-05-10
No UnknownHostException is thrown when invalid host is used. [JIRA: CLIENTS-401]


calvernaz:
Given an invalid host I was excepting an exception has the documentation states.

After a bit of debugging, it seems the excepted exception it's being swallowed 
and we should rely on _isUnresolved_ method.

**Testing code:**

``` java
client = RiakClient.newClient(Lists.newArrayList("no_place_like_home.com"));
```

**Riak client handling**

``` java
bootstrap.handler(new RiakChannelInitializer(this))
     .remoteAddress(new InetSocketAddress(remoteAddress, port));
```

**InetSocketAddress handling**

``` java
    public InetSocketAddress(String hostname, int port) {
        checkHost(hostname);
        InetAddress addr = null;
        String host = null;
        try {
            addr = InetAddress.getByName(hostname);
        } catch(UnknownHostException e) {
            host = hostname;
        }
        holder = new InetSocketAddressHolder(host, addr, checkPort(port));
    }
```

I was wondering if we could fix this behaviour or if there is any reason for that, like in a situation of many nodes and just one is not possible to resolve.

**Version details:**

``` text
$ java -version
java version "1.7.0_13"
Java(TM) SE Runtime Environment (build 1.7.0_13-b20)
Java HotSpot(TM) 64-Bit Server VM (build 23.7-b01, mixed mode)
```

``` xml
<dependency>
  <groupId>com.basho.riak</groupId>
  <artifactId>riak-client</artifactId>
  <version>2.0.0</version>
</dependency>
```

alexmoore:
Thanks for the submission!

alexmoore:
This was merged on the develop branch, and should be out with the next release (today or next week).

PR #519, opened 2015-05-13
CLIENTS-401 - Throw exception when resolving host is not possible

calvernaz:
A possible fix for the issue https://github.com/basho/riak-java-client/issues/518


alexmoore:
Had to chew on this one but I think you've got it correct here. We should throw it on cluster.addNode() but not cluster.start(), with logging on the later.  

With RiakCluster, it should follow a "best-effort" path w.r.t. multiple nodes (healthcheck should mark misconfigured nodes as bad), so we should alert with the log and move along as much as possible.

PR #525, opened 2015-07-08
Ready: Jira 301 - Temporary Queue for operations when RiakNode isn't able to get a connection

christophermancini:
Added ability to queue operations when max connections have been reached to be executed when the next connection is released.

Added gitattributes file to enforce LF line separators and converted files with CRLF line separators to LF.

Please note that I added a new test to RiakNodeTest.


alexmoore:
I'm also worried that if we have 1 thread that keeps getting all the work off the queue, that it could get to a Stack Overflow point.

christophermancini:
Code looks good and tests pass.

christophermancini:
:+1: 255fac4

Issue #527, opened 2015-07-17
Source is using JDK7 items, when target is 1.6 [JIRA: CLIENTS-521]


jplock:
Currently in the pom, the maven-compiler-plugin says the source and target are `1.6`, however Eclipse complains about two usages of JDK7 code in the codebase:
- https://github.com/basho/riak-java-client/blob/develop/src/main/java/com/basho/riak/client/api/RiakClient.java#L265 - `getHostString() is not visible`
- https://github.com/basho/riak-java-client/blob/develop/src/test/java/com/basho/riak/client/core/fixture/NetworkTestFixture.java#L20 - `java.net.StandardSocketOptions` was introduced in JDK7

I guess we could fix this by either increasing the minimum version to `1.7` or find JDK6 equivalents.

alexmoore:
Hey @jplock, not sure how this slipped past - we have the animal sniffer plugin enabled but something's not obviously working.

alexmoore:
Ok, this was fixed on the the above PR, and should be out with the next release (today or next week).

PR #530, opened 2015-07-23
Fixing call to JDK7+ method to maintain JDK6 Runtime compatibility

alexmoore:
Fix for #527 (CLIENTS-521).
- Switched out https://github.com/basho/riak-java-client/blob/develop/src/main/java/com/basho/riak/client/api/RiakClient.java#L265 - `getHostString()` for `getHostName()`, which should be in JRE6.
- https://github.com/basho/riak-java-client/blob/develop/src/test/java/com/basho/riak/client/core/fixture/NetworkTestFixture.java#L20 - `java.net.StandardSocketOptions` was introduced in JDK7, and remains as it's within a test and shouldn't be a runtime issue.

### DIFF
--- a/src/main/java/com/basho/riak/client/api/RiakClient.java
+++ b/src/main/java/com/basho/riak/client/api/RiakClient.java
@@ -262,7 +262,7 @@ public class RiakClient
         List<RiakNode> nodes = new LinkedList<RiakNode>();
         for (InetSocketAddress addy : addresses)
         {
-            builder.withRemoteAddress(addy.getHostString())
+            builder.withRemoteAddress(addy.getHostName())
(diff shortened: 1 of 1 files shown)


christophermancini:
:+1: be100a0

jplock:
Related to this, are there plans to make 1.7 the default since 1.6 was end of life on Feb. 2013.  JDK7 was end of life on April 2015 but I realize most enterprises are still using JDK7 at this point.  Removing JDK6 would at least encourage people to start moving toward something more recent.

http://www.oracle.com/technetwork/java/eol-135779.html

alexmoore:
I know :(  I'll have to double check that there aren't any customers still on 1.6, I'd love to advance it as well.

PR #531, opened 2015-07-28
Jira 453

christophermancini:
Add streaming support to list buckets.


christophermancini:
I still need to get the test working and add more docs.

PR #537, opened 2015-08-26
Not Ready: Updates to Java Client for Data Platform work.

alexmoore:


alexmoore:
We still need the updates to riak_pb before we can merge & test.

alexmoore:
As per @srgg, this is no longer needed.

mdigan:
@alexmoore @srgg this PR was probably super stale, but we still need to get the parallel extract/coverage plan stuff merged at some point. Or is that in a different PR now?

alexmoore:
@mdigan Good catch, we should do a new PR that's not so stale based on the latest merges that @srgg has done.

PR #541, opened 2015-08-27
The RiakIndex annotation should not have a default value [JIRA: CLIENTS-559]

empovit:
Removed the default value from the RiakIndex annotation because it requires a value. Updated the unit tests accordingly to test if a value provided by the user is not an empty string.


alexmoore:
+1, thanks for the PR!

PR #544, opened 2015-09-04
Introduce generic PB Future operation

srgg:
It seems that it is a good time to introduce a generic class that takes care about decoding and encoding to/from RiakMessage


alexmoore:
Very nice, we probably need this in more places.

PR #550, opened 2015-09-23
READY: Settable Default Charset

alexmoore:
This is from a customer request (JIRA CLIENTS-547), to have a settable default charset. Currently it's set to the JRE's default, which can change from machine to machine.   

You can set it from the static [`set()`](https://github.com/basho/riak-java-client/pull/550/files#diff-4a31b593bfc4aa074e9f6cb60083dc5eR94) method on the `DefaultCharset` class, or from the command line by providing the charset name with the arg `com.basho.riak.client.DefaultCharset`. (e.g. `java -Dcom.basho.riak.client.DefaultCharset="UTF-8" ...`)

Rendered Javadocs:
defaultcharset__riak_client_for_java_2_0_3-snapshot_api_

### DIFF
--- a/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
+++ b/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
@@ -8,7 +8,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * <p>
- *     Holds an application-wide default charset, that
+ *     Holds an classloader-wide default charset, that
  *     is then used to encode/decode between Strings and
  *     Byte arrays for Riak's use (see {@link com.basho.riak.client.core.util.BinaryValue}).
  *</p>
@@ -79,8 +79,8 @@ public final class DefaultCharset
     }
 
     /**
-     * Get the current application-wide default Charset for the Riak client.
-     * @return The current application-wide default charset.
+     * Get the current classloader-wide default Charset for the Riak client.
+     * @return The current classloader-wide default charset.
      */
     public static Charset get()
     {
@@ -88,8 +88,8 @@ public final class DefaultCharset
     }
 
     /**
-     * Set the application-wide default Charset for the Riak client.
-     * @param charset The charset to set the application-wide default to.
+     * Set the classloader-wide default Charset for the Riak client.
+     * @param charset The charset to set the classloader-wide default to.
      */
     public static void set(Charset charset)
     {
--- a/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
+++ b/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
@@ -56,26 +56,37 @@ public final class DefaultCharset
 
     private DefaultCharset(Charset c)
     {
-        LogCharsetChange(c);
+        logger.info("Initializing client charset to: {}", c.name());
         this.currentCharset = new AtomicReference<Charset>(c);
     }
 
     private static DefaultCharset initializeDefaultCharsetSingleton()
     {
-        Charset charset = Charset.defaultCharset();
+        Charset charset;
+        final Charset systemDefault = Charset.defaultCharset();
+        final String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
 
-        String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
         if(declaredCharsetName != null && !declaredCharsetName.isEmpty())
         {
-            charset = Charset.forName(declaredCharsetName);
+            try
+            {
+                charset = Charset.forName(declaredCharsetName);
+            }
+            catch (Exception ex)
+            {
+                charset = systemDefault;
+                logger.warn("Requested charset '{}' is not available, the default charset '{}' will be used",
+                            declaredCharsetName,
+                            charset.name());
+            }
+        }
+        else
+        {
+            logger.info("No declared charset found, the default charset '{}' will be used", systemDefault.name());
+            charset = systemDefault;
         }
 
         return new DefaultCharset(charset);
-    }
-
-    private static void LogCharsetChange(Charset charset)
-    {
-        logger.info("Setting client charset to: {}", charset.name());
     }
 
     /**
@@ -93,7 +104,8 @@ public final class DefaultCharset
      */
     public static void set(Charset charset)
     {
-        LogCharsetChange(charset);
+        final Charset current = instance.currentCharset.get();
+        logger.info("Setting client charset from '{}' to '{}'", current.name(), charset.name());
         instance.currentCharset.set(charset);
     }
 }
--- a/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
+++ b/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
@@ -1,0 +1,43 @@
+package com.basho.riak.client.core.util;
+
+import java.nio.charset.Charset;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ *
+ */
+public final class DefaultCharset
+{
+    private final AtomicReference<Charset> currentCharset;
+    private static DefaultCharset instance = initializeDefaultCharsetSingleton();
+
+    private static DefaultCharset initializeDefaultCharsetSingleton()
+    {
+        Charset charset = Charset.defaultCharset();
+
+        String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
+        if(declaredCharsetName != null && !declaredCharsetName.isEmpty())
+        {
+            Charset declaredCharset = Charset.forName(declaredCharsetName);
+            charset = declaredCharset;
+        }
+
+        return new DefaultCharset(charset);
+    }
+
+    private DefaultCharset(Charset c)
+    {
+        this.currentCharset = new AtomicReference<Charset>(c);
+    }
+
+
+    public static Charset get()
+    {
+        return instance.currentCharset.get();
+    }
+
+    public static void set(Charset charset)
+    {
+        instance.currentCharset.set(charset);
+    }
+}
--- a/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
+++ b/src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
@@ -1,5 +1,8 @@
 package com.basho.riak.client.core.util;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.nio.charset.Charset;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -8,8 +11,16 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public final class DefaultCharset
 {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultCharset.class);
+    private final static DefaultCharset instance = initializeDefaultCharsetSingleton();
+
     private final AtomicReference<Charset> currentCharset;
-    private static DefaultCharset instance = initializeDefaultCharsetSingleton();
+
+    private DefaultCharset(Charset c)
+    {
+        LogCharsetChange(c);
+        this.currentCharset = new AtomicReference<Charset>(c);
+    }
 
     private static DefaultCharset initializeDefaultCharsetSingleton()
     {
@@ -25,11 +36,10 @@ public final class DefaultCharset
         return new DefaultCharset(charset);
     }
 
-    private DefaultCharset(Charset c)
+    private static void LogCharsetChange(Charset charset)
     {
-        this.currentCharset = new AtomicReference<Charset>(c);
+        logger.info("Setting client charset to: {}", charset.name());
     }
-
 
     public static Charset get()
     {
@@ -38,6 +48,7 @@ public final class DefaultCharset
 
     public static void set(Charset charset)
     {
+        LogCharsetChange(charset);
         instance.currentCharset.set(charset);
     }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,99 @@
...
+        LogCharsetChange(c);
+        this.currentCharset = new AtomicReference<Charset>(c);
+    }
+
+    private static DefaultCharset initializeDefaultCharsetSingleton()
+    {
+        Charset charset = Charset.defaultCharset();
+
+        String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
+        if(declaredCharsetName != null && !declaredCharsetName.isEmpty())
```
lukebakken: `if (`

This happens in several places.

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,99 @@
...
+ * <p>
+ *     With this class you may change the Riak client to use a different default.
+ *     You can set it at startup by providing the desired Charset name with the vm argument
+ *     {@code -Dcom.basho.riak.client.DefaultCharset="UTF-8"}, or at runtime
+ *     with the static method (see {@link #set(Charset)}).
+ * </p>
+ *
+ * <p>
+ *     As of 2.0.3 it still defaults to the value provided by
+ *     {@link Charset#defaultCharset()},
```
lukebakken: Why would we not always want to default to the JVM default?

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,99 @@
...
+ *     after application startup once you upgrade</b>.
+ * </p>
+ * @author Alex Moore <amoore AT basho dot com>
+ * @since 2.0.3
+ */
+
+public final class DefaultCharset
+{
+    private static final Logger logger = LoggerFactory.getLogger(DefaultCharset.class);
+    private final static DefaultCharset instance = initializeDefaultCharsetSingleton();
```
lukebakken, line 54: I don't know the specifics of JVM instantiation vs C#, but is this guaranteed to be thread safe? Do we care?

In C# I prefer the nested class approach and it appears to be a good way of doing it in Java too:

http://literatejava.com/jvm/fastest-threadsafe-singleton-jvm/

srgg, line 54: The most simple way, but not the most obvious one,  is to declare it as an enum https://dzone.com/articles/singleton-design-pattern-–

THere is also interesting academic case with hacking the Singleton...  For the enum approach there is no chance to hack it with Reflection.

lukebakken, line 54: That is both interesting and gross :-)

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,99 @@
...
+        String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
+        if(declaredCharsetName != null && !declaredCharsetName.isEmpty())
+        {
+            charset = Charset.forName(declaredCharsetName);
+        }
+
+        return new DefaultCharset(charset);
+    }
+
+    private static void LogCharsetChange(Charset charset)
```
srgg: Better to move logging to the set() and use set() from the constructor

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private DefaultCharset(Charset c)
+    {
+        LogCharsetChange(c);
+        this.currentCharset = new AtomicReference<Charset>(c);
+    }
+
+    private static DefaultCharset initializeDefaultCharsetSingleton()
+    {
+        Charset charset = Charset.defaultCharset();
```
srgg: more natural is to reimplement with blank final and introduce a fallback charset:

``` java
final charset;
final String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
if(declaredCharsetName != null && !declaredCharsetName.isEmpty()){
    try {
      charset = Charset.forName(declaredCharsetName);    
    } catch (Exception ex){
      charset = Charset.defaultCharset();
      logger.warn("Requested charset '' is not available, the default charset '' will be used");
    }
} else {
  charset = Charset.defaultCharset();
}
```

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
```diff
@@ -0,0 +1,49 @@
+package com.basho.riak.client.core.util;
+
+import java.nio.charset.Charset;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+public class DefaultCharsetTest
+{
```
srgg, line 16: It might be a good idea to fix behavior for the case when unavailable charset is requested(set during runtime)

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
alexmoore, line 54: The static initializer _should_ only run once, the JVM uses a lock the first time a class is loaded: http://docs.oracle.com/javase/specs/jls/se7/html/jls-12.html#jls-12.4.2.  Also it should be initialized the first time the `get()` method is called: http://docs.oracle.com/javase/specs/jls/se7/html/jls-12.html#jls-12.4.1. 

The AtomicReference should take care of any other cross-threading fun.

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
alexmoore: It can vary from machine to machine, and from dev to server environments.  We had a customer complain that stuff wasn't getting encoded with UTF-8 by default, which would be the logical choice nowadays.

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
alexmoore: :+1: with 9474a5b.

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
alexmoore, line 16: :+1: with 9474a5b.

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
alexmoore: :+1: with 9474a5b. (mostly). Don't know if we can use set from constructor since it needs the instance to be set, in order to set the atomicref.

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
alexmoore: What happens several places? I only see one `if` :)

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,111 @@
...
+        this.currentCharset = new AtomicReference<Charset>(c);
+    }
+
+    private static DefaultCharset initializeDefaultCharsetSingleton()
+    {
+        Charset charset;
+        final Charset systemDefault = Charset.defaultCharset();
+        final String declaredCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
+
+        if(declaredCharsetName != null && !declaredCharsetName.isEmpty())
```
lukebakken: :cry: `if (`

src/main/java/com/basho/riak/client/core/util/DefaultCharset.java
```diff
@@ -0,0 +1,111 @@
...
+            {
+                charset = systemDefault;
+                logger.warn("Requested charset '{}' is not available, the default charset '{}' will be used",
+                            declaredCharsetName,
+                            charset.name());
+            }
+        }
+        else
+        {
+            logger.info("No declared charset found, the default charset '{}' will be used", systemDefault.name());
```
srgg: IMO it may confuse the user, it worth to use less confusing message like:
   Desired charset is not specified, the default charset '{}' will be used

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
srgg, line 16: Excuse me for misleading comment, my point was that it may worth to create corresponding test case to fix the desired behavior in it. If test passed, we have an expected behavior. Kind of 
void setSystemDefaultCharset_IF_desired_is_unavailable()

alexmoore: Ahh ok. Yeah, we should have that test here: https://github.com/basho/riak-java-client/pull/550/files#diff-09fa926e842365e4bcfc9a7cd97443f8R18

alexmoore, line 16: At runtime we only let them set it (using `DefaultCharset.set()`) from an existing charset, so it would have to exist / be available then.  Are you talking about a different case?

srgg, line 16: As for Runtime, you are totally right, but what about the case when wrong charset specified via Constants.CLIENT_OPTION_CHARSET. https://github.com/basho/riak-java-client/pull/550#discussion-diff-40281831R78

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
```diff
@@ -0,0 +1,152 @@
...
+            staticInitializer.setAccessible(true);
+
+            // Set system property for client charset to bad value
+            System.setProperty(Constants.CLIENT_OPTION_CHARSET, goodCharsetName);
+
+            // Run static initializer to test good charset handling
+            DefaultCharset initializedDefaultCharset = (DefaultCharset) staticInitializer.invoke(null);
+
+            // Get initialized charset
+            Field currentCharsetField = DefaultCharset.class.getDeclaredField("currentCharset");
```
srgg: As we want to freeze the observable behavior, here is no needs to check internal state. better to verify as:

```
        assertEquals(systemDefault.name(), DefaultCharset.get().name());
```

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+            else
+            {
+                System.clearProperty(Constants.CLIENT_OPTION_CHARSET);
+            }
+        }
+    }
+
+    @Test
+    public void testStaticInitializerInvalidDefinedCharset()
```
srgg: It might be a good idea to use kind of a [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development) name for the test cases which are intended to freeze a  behavior.  Such approach will shed a bit more light on the expected behavior. For this case it might be something like:

```
testTheDefaultSytemCharsetShouldBeUsedIfTheWrongOneWasRequested()
```

or something like this.

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
```diff
@@ -0,0 +1,152 @@
...
+            currentCharsetField.setAccessible(true);
+            AtomicReference<Charset> initializedCharset =
+                    (AtomicReference<Charset>)currentCharsetField.get(initializedDefaultCharset);
+
+            assertEquals(goodCharset.name(), initializedCharset.get().name());
+        }
+        finally
+        {
+            // Cleanup
+            if (definedCharsetName != null)
```
srgg: To avoid copypasting this block might be moved into tearDown method.

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
```diff
@@ -0,0 +1,152 @@
...
+        Charset definedCharset = Charset.forName(definedCharsetName);
+        assertEquals(definedCharset, clientDefault);
+    }
+
+    @Test
+    public void testStaticInitializerValidDefinedCharset()
+            throws NoSuchMethodException, InvocationTargetException, IllegalAccessException, NoSuchFieldException
+    {
+        String definedCharsetName = System.getProperty(Constants.CLIENT_OPTION_CHARSET);
+        Charset systemDefault = Charset.defaultCharset();
```
srgg: This var (systemDefault) might be refactored to class final static protected attribute and reused after that whenever it needs. IMO  It will increase readability.

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
```diff
@@ -0,0 +1,152 @@
...
+        String badCharsetName = "Foobar-16NE";
+
+        try
+        {
+            // Open up the private static initializer
+            Method staticInitializer = DefaultCharset.class.getDeclaredMethod("initializeDefaultCharsetSingleton",
```
srgg: To avoid copypasting the utility methos might be introduced:

``` java
private void reInitializeDefaultCharsetWith(String desiredCharset)
```

src/test/java/com/basho/riak/client/core/util/DefaultCharsetTest.java
alexmoore: But we're just testing/calling the initializer method b/c the charset for that class is static final, so we can't change it unless we restart the VM. Testing the initializer function should test everything except the actual assignment to the static final "singleton" variable.

PR #561, opened 2015-10-28
Ready: Remove sets / maps / float32s from timeseries

alexmoore:
Covers RTS-483 + RTS-500.
We got ahead of ourselves, we don't support these yet.


src/main/java/com/basho/riak/client/core/converters/TimeSeriesPBConverter.java
```diff
@@ -72,26 +85,28 @@ public static QueryResult convertPbQueryResp(RiakKvPB.TsQueryResp response)
 
     public static Collection<RiakKvPB.TsRow> convertRowsToPb(Collection<Row> rows)
     {
-        ArrayList<RiakKvPB.TsRow> pbRows = new ArrayList<RiakKvPB.TsRow>(rows.size());
+        final ArrayList<RiakKvPB.TsRow> pbRows = new ArrayList<RiakKvPB.TsRow>(rows.size());
 
         for (Row row : rows)
         {
-            pbRows.add(convertRowToPb(row));
+            final RiakKvPB.TsRow.Builder rowBuilder = RiakKvPB.TsRow.newBuilder();
```
srgg, line 68: It will be good for memory consumption if we have re-usable builder

alexmoore, line 68: Looks like we may be able to do that with the clear message. 

https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Message.Builder#clear()

Just need to make it thread safe :)

alexmoore:
I'd say for time's sake let's leave that out for 1.0, we can put it in the next version.

src/main/java/com/basho/riak/client/core/converters/TimeSeriesPBConverter.java
alexmoore, line 68: I'd say for time's sake, let's put this in vNext.

src/main/java/com/basho/riak/client/core/converters/TimeSeriesPBConverter.java
```diff
@@ -203,10 +202,10 @@ else if(columnType == ColumnDescription.ColumnType.SET) // Set
 
     private static ColumnDescription convertPBColumnDescription(RiakKvPB.TsColumnDescription pbColumn)
     {
-        String name = pbColumn.getName().toStringUtf8();
+        final String name = pbColumn.getName().toStringUtf8();
 
-        ColumnDescription.ColumnType type = ColumnDescription.ColumnType.valueOf(pbColumn.getType().getNumber());
-        List<ColumnDescription.ColumnType> complexType =
+        final ColumnDescription.ColumnType type = ColumnDescription.ColumnType.valueOf(pbColumn.getType().getNumber());
+        final List<ColumnDescription.ColumnType> complexType =
```
srgg: As we are removing Map support, shall we also remove complexTypes?

alexmoore: Complex types was another future proofiing, I'll remove it for now as it's unused, as well as the set/map column type declarations.

src/main/java/com/basho/riak/client/core/query/timeseries/ColumnDescription.java
```diff
@@ -43,9 +38,7 @@ public ColumnType getType()
         INTEGER(1),
         FLOAT(2),
```
srgg, line 39: To avoid confusion better to rename FLOAT to DOUBLE and INTEGER to LONG

alexmoore, line 39: This one is a little more interesting, since long is an even more loaded word. Are we referring to C's  32-bit `long`s, the 64-bit on 32-bit arch `long long`s, etc.  If it was SQL Server we would have `bigint`, Oracle would be `Numeric(someBigNum)`...

srgg, line 39: You are right, we need to think about corresponding "names"

srgg:
Everything except 2 weird ColumnTypes (INTEGER, FLOAT) looks great for me

alexmoore:
Ok, those should be addressed in another PR coming soon.

PR #569, opened 2015-11-05
Fixing messagecode output in exception to match PB spec

alexmoore:
Java's bytes are signed, our PB message code byte is not. This fix correctly outputs the correct message code value when creating a "wrong message received" exception:

"Wrong response; expected 93 received -1"  
should be 
"Wrong response; expected 93 received 255"


christophermancini:
:+1: Looks good. Tests pass.

PR #577, opened 2015-12-03
Simplify Riak cluster creation HostAndPort was added

srgg:


src/main/java/com/basho/riak/client/core/util/HostAndPort.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.io.Serializable;
+import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An immutable representation of a host and port.
+ *
+ * This Own implementation is created to reduce the amount of 3rd party libraries
```
alexmoore: What's "Own" ?

src/main/java/com/basho/riak/client/core/util/HostAndPort.java
```diff
@@ -0,0 +1,139 @@
...
+    private final int port;
+    private final String host;
+    private transient InetSocketAddress inetAddress;
+
+    private HostAndPort(String host, int port){
+        this.host = host;
+        this.port = port;
+    }
+
+    public boolean hasPort() {
```
alexmoore: Let's move the brackets to the next line style here. 

```
if (blah)
{...
```

alexmoore:
Looks mostly good, just need to fix the formatting of the brackets to match the project.  +0.75 otherwise.

srgg:
@alexmoore I also realized that introduced approach should be propagated to the RiakClient level as a newClient() routine.. Will add it and adjust PR according your comments.

PR #589, opened 2016-01-06
RTS 795

alexmoore:
Built off the work from RTS-665, this adds a fancier interface to the describe command.

### DIFF
--- a/src/main/java/com/basho/riak/client/core/query/timeseries/CollectionConverters.java
+++ b/src/main/java/com/basho/riak/client/core/query/timeseries/CollectionConverters.java
@@ -84,9 +84,9 @@ public final class CollectionConverters
         final String name = cells.get(0).getVarcharAsUTF8String();
         final String typeString = cells.get(1).getVarcharAsUTF8String();
         final boolean isNullable = cells.get(2).getBoolean();
-        final boolean isPrimaryKeyMember = cells.get(3) != null;
+        final boolean isPartitionKeyMember = cells.get(3) != null;
         final boolean isLocalKeyMember = cells.get(4) != null;
-        final Integer primaryKeyOrdinal = isPrimaryKeyMember ? new Long(cells.get(3).getLong()).intValue() : null;
+        final Integer partitionKeyOrdinal = isPartitionKeyMember ? new Long(cells.get(3).getLong()).intValue() : null;
         final Integer localKeyOrdinal = isLocalKeyMember ? new Long(cells.get(4).getLong()).intValue() : null;
 
         final ColumnDescription.ColumnType type =
@@ -95,7 +95,7 @@ public final class CollectionConverters
         return new FullColumnDescription(name,
                                          type,
                                          isNullable,
-                                         primaryKeyOrdinal,
+                                         partitionKeyOrdinal,
                                          localKeyOrdinal);
     }
 }
--- a/src/main/java/com/basho/riak/client/core/query/timeseries/CollectionConverters.java
+++ b/src/main/java/com/basho/riak/client/core/query/timeseries/CollectionConverters.java
@@ -64,4 +64,38 @@ public final class CollectionConverters
 
         return new ColumnDescription(name, type);
     }
+
+    public static List<FullColumnDescription> convertDescribeQueryResultToColumnDescriptions(QueryResult queryResult)
+    {
+        final List<FullColumnDescription> fullColumnDescriptions = new ArrayList<>(queryResult.getRowsCount());
+
+        for (Row row : queryResult)
+        {
+            fullColumnDescriptions.add(convertDescribeResultRowToFullColumnDescription(row));
+        }
+
+        return fullColumnDescriptions;
+    }
+
+    private static FullColumnDescription convertDescribeResultRowToFullColumnDescription(Row row)
+    {
+        final List<Cell> cells = row.getCellsCopy();
+
+        final String name = cells.get(0).getVarcharAsUTF8String();
+        final String typeString = cells.get(1).getVarcharAsUTF8String();
+        final boolean isNullable = cells.get(2).getBoolean();
+        final boolean isPrimaryKeyMember = cells.get(3) != null;
+        final boolean isLocalKeyMember = cells.get(4) != null;
+        final Integer primaryKeyOrdinal = isPrimaryKeyMember ? new Long(cells.get(3).getLong()).intValue() : null;
+        final Integer localKeyOrdinal = isLocalKeyMember ? new Long(cells.get(4).getLong()).intValue() : null;
+
+        final ColumnDescription.ColumnType type =
+                ColumnDescription.ColumnType.valueOf(typeString.toUpperCase(Locale.ENGLISH));
+
+        return new FullColumnDescription(name,
+                                         type,
+                                         isNullable,
+                                         primaryKeyOrdinal,
+                                         localKeyOrdinal);
+    }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Holds a complete definition for a Table Column in Time Series Riak.
+ * Immutable once created.
+ *
+ * @author Alex Moore <amoore at basho dot com>
+ * @since 2.0.4
+ */
+public class FullColumnDescription
+{
+    private final ColumnDescription baseColumnDescription;
```
srgg: It might be more logical to use inheritance  instead of aggregation:

class FullColumnDescription extends ColumnDescription

src/main/java/com/basho/riak/client/core/query/timeseries/CollectionConverters.java
```diff
@@ -64,4 +64,38 @@ private static ColumnDescription convertPBColumnDescription(RiakTsPB.TsColumnDes
...
+        }
+
+        return fullColumnDescriptions;
+    }
+
+    private static FullColumnDescription convertDescribeResultRowToFullColumnDescription(Row row)
+    {
+        final List<Cell> cells = row.getCellsCopy();
+
+        final String name = cells.get(0).getVarcharAsUTF8String();
```
srgg, line 96: Better to add kinda sanity check of the row, just in case.. Maybe check number of cells or something like this.. As well as add a comment describes expected format

src/main/java/com/basho/riak/client/core/query/timeseries/TableDefinition.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.List;
+
+/**
+ * Holds a definition for a Table in Time Series Riak.
+ * Immutable once created.
+ * @author Alex Moore <amoore at basho dot com>
+ * @since 2.0.4
+ */
+public class TableDefinition
```
srgg, line 11: Might be useful to add  the following utilitarian accessors for getting:
- List of Partition Keys in proper order
- List of Local Keys in proper order
- Description by Column Name

src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
```diff
@@ -0,0 +1,180 @@
...
+
+        if (isNullable != that.isNullable)
+        {
+            return false;
+        }
+        if (!baseColumnDescription.equals(that.baseColumnDescription))
+        {
+            return false;
+        }
+        if (partitionKeyOrdinal != null ?
```
srgg: Not sure whether we should differentiate descriptions that have the same Field Name by the order in local or partition key. We need to think twice about this

src/main/java/com/basho/riak/client/core/query/timeseries/TableDefinition.java
```diff
@@ -0,0 +1,85 @@
...
+{
+    private final String tableName;
+    private final List<FullColumnDescription> fullColumnDescriptions;
```
srgg: It might be a good idea to avoid field name duplication on this level, we have two options:
- Change FullColumnDescription hashCode/equals to use only Field Name and store descriptions in HashSet
- Store descriptions in a HashMap and use FieldName as a key

src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
alexmoore: I started out that way, but since we don't have any cross-use cases between the two I think it's better to keep them separate.  If we want to add inheritance later we can without breaking stuff.  If we ever wanted to remove inheritance though we'd be out of luck. :)

src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
alexmoore: They would be different if you're comparing two rows from different tables and one is a key member, and the other isn't.

srgg: How about the case when 2 fields have identical description despite the fact that they are belong to the different tables?

src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
srgg: What about passing them as an argument:

 // somewhere in a customer's code
  void takimgCareAboutTheSchema(Iterable< ColumnDescription > descriptions)

How can I pass FullColumnDescription to this ^^ routine?

src/main/java/com/basho/riak/client/core/query/timeseries/TableDefinition.java
alexmoore, line 11: What would the use cases for these be?  Spark connector?

alexmoore:
@srgg I think this covers the issues we talked about.

src/main/java/com/basho/riak/client/core/query/timeseries/TableDefinition.java
```diff
@@ -0,0 +1,163 @@
...
+ * Holds a definition for a Table in Time Series Riak.
+ * Immutable once created.
+ * @author Alex Moore <amoore at basho dot com>
+ * @since 2.0.4
+ */
+public class TableDefinition
+{
+    private final String tableName;
+    private final LinkedHashMap<String, FullColumnDescription> fullColumnDescriptions = new LinkedHashMap<>();
+    private final ArrayList<FullColumnDescription> partitionKeys = new ArrayList<>();
```
srgg, line 15: Not sure that we need to pre-cache keys stuff, since not all users need to use it.

IMO, it is better to introduce utilitarian assessors which will return newly created collections each time. As a result caching will be performed on the user side, if the user really needs this.

src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
```diff
@@ -0,0 +1,138 @@
+package com.basho.riak.client.core.query.timeseries;
+
+/**
+ * Holds a complete definition for a Table Column in Time Series Riak.
+ * Immutable once created.
+ *
+ * @author Alex Moore <amoore at basho dot com>
+ * @since 2.0.4
+ */
+public class FullColumnDescription
```
srgg: There is still nothing common between FullColumnDescription and ColumnDescription. What about introducing IColumnDefinition and implement it by both of them?

alexmoore: But then we have a chicken and egg problem with the ColumnType Enum. We'd have to move it elsewhere (concrete enum referenced in interface), which would break anything that references it currently (external or internal).

src/main/java/com/basho/riak/client/core/query/timeseries/TableDefinition.java
alexmoore, line 15: I lean the other way, we can easily create it while in the constructor, and then we have an entire immutable object once we're done.  If we do it later then we're saving 8? object references, and introducing more code and possible race conditions later.

src/main/java/com/basho/riak/client/core/query/timeseries/FullColumnDescription.java
srgg: As for me it is better to move it to the interface, since we haven't published this yet.

src/main/java/com/basho/riak/client/core/query/timeseries/TableDefinition.java
srgg, line 15: Since the whole instance is immutable from the user perspective, there shouldn't be any race conditions.  On the other hand,  as we are not expecting that user will execute describe table each second, I think that we can leave this as is.

srgg:
Looks good to me

alexmoore:
Verbal :+1: from @srgg

PR #590, opened 2016-01-25
Add location accessor

dbyron0:
From the conversation here: http://article.gmane.org/gmane.comp.db.riak.user/17142.  I did a bit of cleanup before adding the actual accessor, though I'd be happy if only the accessor goes in.

Thanks.

-DB

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -266,7 +266,10 @@
                 <configuration>
                     <source>1.7</source>
                     <target>1.7</target>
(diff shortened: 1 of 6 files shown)


alexmoore:
Thanks for the PR!  I put in a test for the location property, and reincluded the SetMemberFilter varargs ctor - don't want to do a major version bump of the client next release.

PR #597, opened 2016-03-05
Filter SSL Ciphers list to those supported by Java and Erlang R16

alexmoore:
Added test to verify this list in the future / on new javas/erlangs. 
Added test to print supported/unsupported lists for future reference.

### DIFF
--- a/src/main/java/com/basho/riak/client/api/RiakClient.java
+++ b/src/main/java/com/basho/riak/client/api/RiakClient.java
@@ -37,18 +37,18 @@ import java.util.concurrent.Future;
  * <img src="doc-files/client-image.png">
  * </p>
  * <p>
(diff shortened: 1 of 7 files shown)


lukebakken:
Just out of curiosity, did anyone run into the `term_to_oid` error with the Java client?

alexmoore:
Not that I know of.

alexmoore:
Since Java seems to correctly negotiate ciphers with Erlang there's nothing to merge anymore, so I'm going to close this PR.  I'll keep the cipher test code around for newer erlangs/javas though instead of deleting the branch.  

I'm also going to keep the `if(supportedProtocols includes TLS1.2) use TLS1.2` code in, Java 7 will use TLS 1.0 by default locally in tests. Can remove it when we drop 7 support.  

So... no code changes right now.

PR #602, opened 2016-03-16
RiakTs CreateTable command

srgg:


src/test/java/com/basho/riak/client/api/commands/CreateTableTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class CreateTableTest extends OperationTestBase<CreateTableOperation>
+{
+    private static final TableDefinition tableDefinition = new TableDefinition("TestTable", Arrays.asList(
+            new FullColumnDescription("field1", ColumnDescription.ColumnType.TIMESTAMP,  false, 1),
+            new FullColumnDescription("field2", ColumnDescription.ColumnType.SINT64, false, null, 1),
```
alexmoore: Do you think this case should result in an error when creating the TableDefinition? (Having 2 columns with the same key ordinal #).  The null ordinal # is confusing too...

srgg: Hm.. my bad.. I forgot about the necessity of having the whole PK as a head of a local key :(

alexmoore:
+1. Works as advertised.  

I'll have to see if we need to add any extra validation to ensure that one key type is null while the other is defined (your second column), but that can wait.

PR #608, opened 2016-04-04
An attempt to unify FutureOperation success and error handling

srgg:
@alexmoore Not sure whether the resulting code is more intuitive than the original code, please feel free to drop it off


alexmoore:
I like the idea of this but it conflicts with the necessary ordering of closing connection -> setException in the error case.  Going to close for now.

PR #621, opened 2016-05-18
Merge the changes we did for TS 1.3 into dpclient branch, before merging back to develop.

alexmoore:
@srgg Here's the changes we need in the dp-client branch before we can merge that one to develop.  

The only test I'm having trouble with at the moment is `ITestFullBucketRead:queryDataByUsingAnAlternateCoveragePlan()`, I'm getting `
java.util.concurrent.ExecutionException: com.basho.riak.client.core.netty.RiakResponseException: primary_partition_unavailable` back.


srgg:
@alexmoore Is it possible to switch between TTB and old PB usage for Fetch/Store/Query operations? Such ability might be very useful for performance testing.

PR #637, opened 2016-07-15
Reduce unused channel close log message to info

### LINKED ISSUE #628: Log "inAvailable channel closed" notification as a lower class [JIRA: CLIENTS-895]
Currently if an unused connection is closed, it gets logged as ERROR.  We should log it as WARNING or INFO.

Example: 

``` java
ERROR [nioEventLoopGroup-2-6:] com.basho.riak.client.core.RiakNode - inAvailable channel closed;
```

alexmoore:
Fixes #628 (CLIENTS-895)

### DIFF
--- a/src/main/java/com/basho/riak/client/core/RiakNode.java
+++ b/src/main/java/com/basho/riak/client/core/RiakNode.java
@@ -136,7 +136,7 @@ public class RiakNode implements RiakResponseListener
                 // the next time it's pulled from the pool.
                 // We record the disco for the health check.
                 recentlyClosed.add(new ChannelWithIdleTime(future.channel()));
-                logger.error("inAvailable channel closed; id:{} {}:{}",
+                logger.warn("Available channel closed; id:{} {}:{}",
                              future.channel().hashCode(), remoteAddress, port);
             }
         };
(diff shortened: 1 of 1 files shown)


src/main/java/com/basho/riak/client/core/RiakNode.java
```diff
@@ -136,7 +136,7 @@ public void operationComplete(ChannelFuture future) throws Exception
                 // the next time it's pulled from the pool.
                 // We record the disco for the health check.
                 recentlyClosed.add(new ChannelWithIdleTime(future.channel()));
-                logger.error("inAvailable channel closed; id:{} {}:{}",
+                logger.warn("Available channel closed; id:{} {}:{}",
```
srgg: @alexmoore Not sure whether it should be a warning, since it is an ordinal situation

alexmoore: It could be normal, or could signify something bigger is going on with the network / application if seen in conjunction with other errors, or many of them grouped together.

https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Level.html#WARN

srgg: IMO it is either INFO or DEBUG

srgg:
👍 LGTM

PR #639, opened 2016-07-15
Remove Support for Java 7

### LINKED ISSUE #635: Ending support for Java 7 [JIRA: CLIENTS-919]
Now that we are[ 1 year past the Java 7 End of Public Updates](http://www.oracle.com/technetwork/java/eol-135779.html), we should end support for it.  Moving to Java 8 will give us some nice code features too.

alexmoore:
Fixes #635 (CLIENTS-919).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -317,8 +317,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.5.1</version>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
                     <compilerArgs>
                         <arg>-Werror</arg>
                         <arg>-Xlint:all</arg>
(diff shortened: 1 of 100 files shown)


jplock:
Are you planning to use Stream's with this as well?  I've spotted a few places which could be converted to streams if needed.

alexmoore:
@jplock I've certainly played with it, but I'm worried about performance.  I'll probably refactor those and do some A/B testing with our performance team to see if it has any noticeable effect.

alexmoore:
@srgg When I add that 'maven-enforcer-plugin' into the code, `mvn install` fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:2.1:shade (default) on project riak-client: Error creating shaded jar: 18607 -> [Help 1]
```

Any idea what's going on there?

alexmoore:
Replacing this PR with #666.

PR #647, opened 2016-07-22
Separate Content-type and charset in RiakObject. 

### LINKED ISSUE #549: Duplicated charset when storing a RiakObject with a content type with charset [JIRA: CLIENTS-602]
If I store a RiakObject with a content type of `application/json;charset=UTF-8`, the charset gets duplicated on the stored value when read by http, ending up with this content type:
`application/json;charset=UTF-8; charset=UTF-8`

I did a small sample to show the issue here: 
https://github.com/gfbett/RiakIssueSample

alexmoore:
Fixes #549 (CLIENTS-602).

Currently the Java client will attach a RiakObject's charset setting to the content-type:

```
RiakObject ro = new RiakObject();
ro.setContentType("application/json");
ro.setCharset("UTF-16");

ro.getContentType(); // -> "application/json;charset=UTF-16"
```

While not 100% wrong, we should separate them.  

If you save the above object to Riak, and fetch it via the HTTP interface you'll get: "application/json;charset=UTF-16; charset=UTF-16" since we duplicate the charset to the proper charset buffer, but don't remove it from the content-type.

This PR introduces a proper `charset` member on the RiakObject class.  We will no longer attach it to the `content-type`, and we don't care if they set one there or not, as to not break existing code. 

If this new charset member is not set it will attempt to see if there is one present in the `content-type` for the `getCharset()` and `hasCharset()` methods, for backwards compatibility.   

Thanks to @gfbett for reporting the issue.

### DIFF
--- a/src/main/java/com/basho/riak/client/core/query/RiakObject.java
+++ b/src/main/java/com/basho/riak/client/core/query/RiakObject.java
@@ -29,13 +29,13 @@ import java.nio.charset.Charset;
  * <p>
  * While the client APIs provide methods to store/retrieve data to/from Riak using your own
  * Beans/POJOs, this class is used as the core data type to and from which those types are
- * converted. 
+ * converted.
  * </p>
  * <p>
  * <h5>Working with Metadata</h5>
  * An object in Riak has a few types of metadata that can be attached; Secondary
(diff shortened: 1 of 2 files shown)


christophermancini:
👍 Code looks good and tests pass.

PR #649, opened 2016-07-26
Merge dpclient rebased with docker

oleksii-suprun:


alexmoore:
@aleksey-suprun Do we need this PR anymore, or is this separate work from the other DP/Spark enhancement PR that was already merged?

srgg:
I don't thunk that it is needed for now

PR #654, opened 2016-08-05
Ready: Add isNotFound parameter to data type responses

### LINKED ISSUE #536: Detect non-existing keys for CRDTs [JIRA: CLIENTS-552]
There is a semantic difference between non-existing key and a default value may me important for downstream consumers. However, CRDT queries return default values: 0 for a counter and empty set for a set. My following examples are in Scala:

```
val fetchOp = new FetchCounter.Builder(location).build()
val c = client.execute(fetchOp).getDatatype // if I get 0, does the key exist in store?
```

For counters, there is a workaround that involves an additional query:

```
val fetchOp = new FetchValue.Builder(location).withOption(FetchValue.Option.HEAD, java.lang.Boolean.TRUE).build()
val res = client.execute(fetchOp) // res.isNotFound should equal(true). 
// Why isNotFound not supported by CRDTs?
```

For sets, there seem to be a hack that a context object should be null if set did not exist before (not sure if it's a valid way though).

The concept of non-existing key is important enough to support it explicitly (like with regular objects in the snippet above).

alexmoore:
Fixes #536 (CLIENTS-552).

### DIFF
--- a/src/main/java/com/basho/riak/client/core/query/crdt/types/RiakCounter.java
+++ b/src/main/java/com/basho/riak/client/core/query/crdt/types/RiakCounter.java
@@ -20,12 +20,12 @@ package com.basho.riak.client.core.query.crdt.types;
  * <p>
  * This is an immutable value returned when querying Riak for a counter datatype.
  * </p>
+ *
  * @author Dave Rusek <drusek at basho dot com>
  * @since 2.0
  */
 public class RiakCounter extends RiakDatatype
 {
-
     private long value = 0;
 
     public RiakCounter(long value)
@@ -35,12 +35,36 @@ public class RiakCounter extends RiakDatatype
 
     /**
      * Get this RiakCounter as a {@link Long} value.
+     *
      * @return a Long value for this RiakCounter.
      */
-	@Override
+    @Override
     public Long view()
     {
         return value;
     }
 
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o)
+        {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass())
+        {
+            return false;
+        }
+
+        RiakCounter that = (RiakCounter) o;
+
+        return value == that.value;
+
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return (int) (value ^ (value >>> 32));
+    }
 }
--- a/src/main/java/com/basho/riak/client/core/query/crdt/types/RiakMap.java
+++ b/src/main/java/com/basho/riak/client/core/query/crdt/types/RiakMap.java
@@ -40,7 +40,6 @@ public class RiakMap extends RiakDatatype
 
     public RiakMap(List<MapEntry> mapEntries)
     {
-
         for (MapEntry entry : mapEntries)
         {
             List<RiakDatatype> datatypes;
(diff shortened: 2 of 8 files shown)


src/main/java/com/basho/riak/client/core/query/crdt/types/RiakMap.java
```diff
@@ -27,37 +27,35 @@
...
+ *
  * @author Dave Rusek <drusek at basho dot com>
  * @since 2.0
  */
 public class RiakMap extends RiakDatatype
 {
-
-    private final Map<BinaryValue, List<RiakDatatype>> entries =
-        new HashMap<BinaryValue, List<RiakDatatype>>();
+    private final Map<BinaryValue, List<RiakDatatype>> entries = new HashMap<>();
```
russelldb, line 39: I don't understand this change, why is it in with a PR that adds a NotFound field to the fetch response?

russelldb:
The code looks good.

One question: there seems to be at least 2 things in this PR 1. the change as described 2. a general refactor.

I think 2 PRs would have been better.

russelldb:
I can't get the iTests to run, I just get `java.lang.AssertionError: Update operation failed: com.basho.riak.client.core.NoNodesAvailableException` even after using java.lang.AssertionError: Update operation failed: com.basho.riak.client.core.NoNodesAvailableException. I can ping the nodes, and `member-status` says they're a cluster and happy.

Will wait for @alexmoore to get in rather than persist at this point

src/main/java/com/basho/riak/client/core/query/crdt/types/RiakMap.java
christophermancini, line 39: Pretty sure this is a resharper code standards format change. He is updating files as he touches them.

christophermancini:
👍 Tests pass, code looks good.

src/main/java/com/basho/riak/client/core/query/crdt/types/RiakMap.java
alexmoore, line 39: Yep, just a general safe refactoring to standardize spacing, formatting etc as I touch files. I should have moved that specific one into a separate commit but probably saw red at the mixing of tabs and spaces.

alexmoore:
@russelldb I'll try to at least split them into different commits in the future.  I've been avoiding a "mega-refactor" PR, just because it would be terrible to review everything all at once.  reformatting as I go seems to work though.

src/main/java/com/basho/riak/client/core/query/crdt/types/RiakCounter.java
```diff
@@ -35,12 +35,36 @@ public RiakCounter(long value)
 
     /**
      * Get this RiakCounter as a {@link Long} value.
+     *
      * @return a Long value for this RiakCounter.
      */
-	@Override
+    @Override
     public Long view()
```
srgg, line 42: Why don't us change Long to long to avoid Boxing..

**UPD:** it turned out that this Long is dropping from RiakDatatype, therefore IMO better to change:

```
 private long value = 0;
```

to use Long instead of long, and maybe even to make RiakDatatype be a Generic class

src/main/java/com/basho/riak/client/core/query/crdt/types/RiakMap.java
```diff
@@ -237,6 +245,30 @@ public RiakRegister getRegister(String key)
...
     }
 
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o)
+        {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass())
```
srgg, line 255: Since this class is not marked as a 'final' it doesn't work for whatever descendant, IMO better to make all CRDT classes final.

Also not sure whether this default 'equals' implementation will be useful in the real life

russelldb:
+1

Tests pass. Looks good. WFM

Issue #656, opened 2016-08-11
Some Tests are getting flappier. [JIRA: CLIENTS-960]


alexmoore:
With the latest TS build, some tests are showing their flappyness more.

RiakClusterFixtureTest.testOperationQueue -> times out
ITestListKeysOperation.testSingleFetch -> fails on key comparison
ITestFullBucketRead -> All tests can throw sibling found exceptions on multi-node test setups.
~~ITestCoveragePlan: fetchAllDataByUsingCoverageContext -> get null pointer exception on [this line](https://github.com/basho/riak-java-client/blob/05fbe31620495feea9f33bc3138132bad0e6cf79/src/test/java/com/basho/riak/client/core/operations/itest/ITestCoveragePlan.java#L157-L157)~~ Moved to issue #658

alexmoore:
Fixed in #657 & #659.

PR #657, opened 2016-08-11
Ready: Make integration tests great again

alexmoore:
Fix for #656 (CLIENTS-960).

c4d8383 Fixes ITestListKeysOperation.testSingleFetch -> fails on key comparison.
414748e Fixes ITestFullBucketRead -> All tests can throw sibling found exceptions on multi-node test setups.

c52e990 + 61f474e  Fixes RiakClusterFixtureTest.testOperationQueue -> times out.

Remainder fix `ITestCoveragePlan: fetchAllDataByUsingCoverageContext -> get null pointer exception`.

### DIFF
--- a/src/main/java/com/basho/riak/client/api/commands/kv/MultiCommand.java
+++ b/src/main/java/com/basho/riak/client/api/commands/kv/MultiCommand.java
@@ -371,9 +371,9 @@ abstract class MultiCommand<BaseCommand extends RiakCommand<BaseResponseType, Lo
         }
 
         @Override
-        public void await(long timeout, TimeUnit unit) throws InterruptedException
+        public boolean await(long timeout, TimeUnit unit) throws InterruptedException
         {
-            latch.await(timeout, unit);
+            return latch.await(timeout, unit);
         }
 
         @Override
--- a/Makefile
+++ b/Makefile
@@ -7,14 +7,16 @@ compile:
 
 test: unit-test integration-test
 
+test-ts: unit-test integration-test-timeseries
+
 unit-test:
 	mvn test
 
 integration-test:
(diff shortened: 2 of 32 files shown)


src/main/java/com/basho/riak/client/api/commands/kv/MultiCommand.java
```diff
@@ -371,9 +371,9 @@ public void await() throws InterruptedException
         }
 
         @Override
-        public void await(long timeout, TimeUnit unit) throws InterruptedException
+        public boolean await(long timeout, TimeUnit unit) throws InterruptedException
         {
```
christophermancini, line 375: Could this cause backwards incompatibility?

christophermancini:
👍 Tests pass. Code looks good. Might want to keep in mind when it comes to release if there are breaking changes in this PR.

src/main/java/com/basho/riak/client/api/commands/kv/MultiCommand.java
alexmoore, line 375: Nope. If they don't assign/use the return value then the behavior is exactly the same.  If they want to use it though, they can change their code.

srgg:
@alexmoore what is this reminder about:

Remainder fix ITestCoveragePlan: fetchAllDataByUsingCoverageContext -> get null pointer exception.

?

alexmoore:
@srgg Oh, I meant the remainder (the rest of) the commits were for that fix.

alexmoore:
@christophermancini Yeah, including a return type where one didn't before shouldn't break anything in this case, we're just passing on a little bit of extra information that has no effect on the computation.

srgg:
👍 LGTM

src/main/java/com/basho/riak/client/api/commands/kv/MultiCommand.java
dbyron0, line 375: According to http://stackoverflow.com/questions/3589946/retrofitting-void-methods-to-return-its-argument-to-facilitate-fluency-breaking/3589948#3589948, this is a source-compatible, but binary-incompatible change.

PR #662, opened 2016-08-23
Disallow 0 as a timeseries timeout value

alexmoore:
Client-side fix for https://github.com/basho/riak_kv/issues/1464.

We were sending a "0" as the default timeout value on TTB TS Fetch requests, should have been the atom `undefined`.   Also disallowing 0 as a valid value in all the TS command / operation builders.


christophermancini:
👍 Tests pass against Riak KV and TS 1.3.1 / 1.4.0. Looks good.

PR #664, opened 2016-08-24
Partially implement withColumns method in the StoreOperation.Builder …

jbrisbin:
Please ensure the following is present in your PR:
- [ ] Unit tests for your change
- [ ] Integration tests (if applicable)

…class to prevent the Spark connector from failing due to calling this method. Understand that this functionality isn't implemented all the way back to Riak. There will need to be some additional work to fully support passing the column names to TS.


alexmoore:
Included in #665 .

PR #677, opened 2016-10-19
Streaming results for List Keys/Buckets, 2i, and MR

alexmoore:
This PR adds support for streaming results in:
- List Keys Command + Operation
- List Buckets Command + Operation
- Secondary Index Commands + Operation
- Map Reduce Commands + Operation
- Full Bucket Read Command (Riak TS only currently)

This was a very interesting exercise in creating a new results return method from top-to-bottom, while not breaking/changing existing APIs. 
- [x] Unit tests for your change
- [X] Integration tests
- [x] Documentation in code
- [X] A docs.basho.com PR for public api changes: basho/basho_docs#2270


src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
```diff
@@ -0,0 +1,122 @@
...
+        this.createNext = createNextFn;
+        this.getNextIterator = getNextIteratorFn;
+        this.getContinuationFn = getContinuationFn;
+        loadNextChunkIterator();
+    }
+
+    @Override
+    public boolean hasNext()
+    {
+        if(currentIteratorHasNext())
```
lukebakken: `if (` :trollface:

src/main/java/com/basho/riak/client/api/commands/ImmediateCoreFutureAdapter.java
```diff
@@ -0,0 +1,83 @@
+/*
```
lukebakken, line 1: These headers should either be nuked or updated.

alexmoore: https://github.com/basho/riak-java-client/issues/678

jbrisbin:
Looks like a a good first step. IMO it would be good to have as an overall goal to make more use of lambdas, not just as callbacks for asynchronous functions but as a tool in the quiver to reduce garbage. If you follow the code path through a request you see a lot of objects being created to support a single operation. Some of that can't be avoided but there are places I think a lambda-centric approach (versus creating a concrete implementation of some abstract class and instantiating it) would be more efficient.

That said, we're probably not at a level where heap usage, GC, and those other considerations are very high priority.

alexmoore:
Indeed, I was lucky to get away with only 4 new class files, and I think I
did a fairly good job about reuse, but there is always room to improve.

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public BinaryValue getContinuation()
+    {
+        return continuation;
+    }
+
+    @Override
+    public synchronized FinalT next()
+    {
+        return createNext.apply(currentIterator.next());
```
srgg: Lack of synchronization on hasNext() whereas next() is synchronized looks looks very suspicious for me.

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
```diff
@@ -0,0 +1,122 @@
...
+        // Core Operation Not Done OR items still in chunk Queue
+        return !coreFuture.isDone() || !chunkQueue.isEmpty();
+    }
+
+    public boolean hasContinuation()
+    {
+        return continuation != null || possibleChunksRemaining();
+    }
+
+    public BinaryValue getContinuation()
```
srgg, line 158: Why don't us get rid of BinaryValue here?

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
```diff
@@ -0,0 +1,138 @@
...
+
+public class ChunkedResponseIterator<FinalT, ChunkT extends Iterable<CoreT>, CoreT> implements Iterator<FinalT>
+{
+    private final int timeout;
+    private volatile BinaryValue continuation = null;
+    private final StreamingRiakFuture<ChunkT, ?> coreFuture;
+    private final TransferQueue<ChunkT> chunkQueue;
+    private final Function<CoreT, FinalT> createNext;
+    private final Function<ChunkT, Iterator<CoreT>> getNextIterator;
+    private final Function<ChunkT, BinaryValue> getContinuationFn;
```
srgg, line 54: Is there a reason why it may require to expose BinaryValue as a part of getContinuationFn() signature?

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
alexmoore: We may not need the synchronization here, as iterators aren't typically guaranteed to be reentrant.  The resulting iterator is single-use only too, so I'll need to make sure I document that somewhere as well.

alexmoore: Originally I did all the lower-level iterator checking/loading in `next()` instead of `hasNext()` like it is now.  I'll just remove it based on my previous comment.

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
alexmoore: Continuation values (if the command/operation uses them) are always returned as BinaryValue through our interface.  The continuation also comes as part of one of the chunked responses, so we need to watch out for it.

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
alexmoore: See above.

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
```diff
@@ -0,0 +1,138 @@
...
+                    populatedChunkLoaded = currentIteratorHasNext();
+
+                    loadContinuation(nextChunk);
+                }
+            }
+        }
+        catch (InterruptedException e)
+        {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
```
srgg: Changing an interruption behavior by re-throwing RuntimeException calling interrupt() in the same time is not a common pattern. I have no clear understanding what we-ve got as a result. Theoretically, the most possible behavior is -- the original InterruptedException will be swallowed.

lukebakken: FWIW, I didn't grok this either and just assumed it was the right way :fried_shrimp:

alexmoore: So what I settled on was to mark the interrupt() flag on the thread and to wrap it in a RuntimeException(), that way the thread should give up control.  

If we get an interrupt here it means that the user thread really wanted to interrupt for whatever reason, and we can't block indefinitely while we want for another chunk.  In the "Riak is being slow", or "the user did a very bad/slow query" case, blocking indefinitely could deadlock the user thread for longer than wanted. 

Also we can't extend `hasNext()` signature to pass along the InterruptedException, hence the wrap.  I need to add a test to see if the user calls `cancel()` if the future/iterator will pick it up and stop iteration.

lukebakken:
This all _seems_ fine to a non-Java dev. While I know why the queue and `.poll()` is used, it seems hacky. Feeble searching for information about netty brought up [this SO post](http://stackoverflow.com/questions/15286846/netty-where-is-the-callback) which talks about methods to execute callbacks on threads that don't block the IO thread. Maybe something to look into for the future?

src/test/java/com/basho/riak/client/api/commands/indexes/itest/ITestIndexBase.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.basho.riak.client.api.commands.indexes.itest;
+
+import com.basho.riak.client.core.operations.itest.ITestBase;
+import com.basho.riak.client.core.query.Namespace;
+import org.junit.AfterClass;
+
+import java.util.concurrent.ExecutionException;
+
+public abstract class ITestIndexBase extends ITestBase
```
srgg, line 25: We already have ITestAutoCleanupBase that do mostly the same

alexmoore: Mostly, but ITestAutoCleanupBase cleans up after every test, while this one cleans up after each suite/class of tests.  With some of the tests doing 1K/10K entries it speeds things up a bit to use common data across all a class's tests. 

I wanted to also put an abstract `setupData()` method in here but I couldn't get the RiakCluster/RiakClient dependency and it's initialization correct with all the static context going on.

alexmoore:
@jbrisbin @srgg One other thing I hoped you pick on was the unbounded queue we use, and the way I pass in the poll timeout from the outside.  What do you think about these two designs?

If we used a bounded queue, it would limit Java memory usage if the user isn't consuming the iterator, but then the network buffers would bulge instead, since Riak just sends data like a firehose. 

With passing the poll timeout, I was thinking we might also be able to do some sort of devolving timeout based on success rate of the last few polls and their timeouts, but not sure how much extra work that would be.

jbrisbin:
@a1 We could have a variant where you pass in the Queue to use, which could be any implementation, including one with blocking publish semantics to back up the TCP end in case the queue is not being emptied fast enough.

The read timeout should still probably be configured externally since that's a consideration that takes place inside our code.

alexmoore:
@jbrisbin Actually now I remember why I made it unbounded. If we block on the Netty thread for any reason it will throw a BlockingOperationException, which will kill the whole operation. Very bad for large streaming ops because it'll then retry, and fail, and retry and fail, and Riak will then have 3 listing operations running. 

Also if we expose a "provide your own Queue", we'll need to expose the type info all the way up (unless it's only exposed at the Operation level, and not the Command level).

alexmoore:
Ok, that should be all the gold plating I can think of.

src/main/java/com/basho/riak/client/api/commands/ChunkedResponseIterator.java
srgg: @a1 The more I think about  the more I feel like there should be enough to throw teh only Runtime exception. We can't throw Interrupted due to the contract, therefore nobody will expect that any of it's methods might be blocked and as result nobody will expect that thread may be marked as interrupted.  If you approve, I will introduce corresponding changes as a separate commit.

alexmoore: Reading more opinions about this... I guess we could add a flag to note that it was interrupted, and keep going. If we get to the end (`hasNext()` returns false) we could do something at that point, but not sure what since the iterator interface won't allow us to interrupt.

srgg: My initial thought was exactly the same: set the flag and move forward without raising any exception. If this call was made from the whatever execution thread the programmer should take  care about the proper thread interuption and at least we do not make it worse.

Please excuse me for my hesitation.

alexmoore:
@srgg FullBucketRead tests pass locally, +1 to 51bff20.  Going to review the rest of it now to see if anything jumps out.

src/main/java/com/basho/riak/client/api/commands/indexes/SecondaryIndexQuery.java
```diff
@@ -686,8 +709,19 @@ public boolean hasEntries()
...
+         * Get a list of the result entries for this response.
+         * If using the streaming API this method will return an empty list.
+         *
+         * @return A list of result entries.
+         */
         public final List<E> getEntries()
         {
+            if(isStreamable())
+            {
+                return new ArrayList<>(0);
```
srgg: What about throwing either IllegalState or UnsupportedOperation?

srgg: if not, it is better to return Collections.EMPTY_LIST

alexmoore (commented):
+1 to any changes I didn't do.   Also added some misc public api javadocs and did some whitespace cleanup.

srgg:
+1 to any changes I didn't do.

lukebakken:
Looks like more code was deleted than written, always a good sign. Tests pass with Riak 2.2.0, 2.1.4, and 2.0.7 (except one YZ test there). 👍

PR #683, opened 2016-10-31
Added AutoCloseable interface to RiakClient

olitazl:
I just added the AutoCloseable Interface to RiakClient to use the Try-with-Resource Statement.

### DIFF
--- a/src/main/java/com/basho/riak/client/api/RiakClient.java
+++ b/src/main/java/com/basho/riak/client/api/RiakClient.java
@@ -31,6 +31,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.lang.AutoCloseable;
 
 /**
  * <script src="https://google-code-prettify.googlecode.com/svn/loader/run_prettify.js"></script>
@@ -158,7 +159,7 @@ import java.util.concurrent.TimeoutException;
  * @author Sergey Galkin <srggal at gmail dot com>
  * @since 2.0
  */
-public class RiakClient
+public class RiakClient implements AutoCloseable
 {
     private final RiakCluster cluster;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/basho/riak/client/api/RiakClient.java
```diff
@@ -416,6 +417,18 @@ public static RiakClient newClient(Collection<HostAndPort> hosts, RiakNode.Build
...
+
+    /**
+     * Implementation of the AutoCloseable-Inteface.
+     * <p>
+     *     The Client will be automatically shut down when used with try-with-resource statement
+     * </p>
+     */
+    @Override
+    public void close(){
+        this.shutdown();
```
srgg: Shutdown method returns Future, therefore to respect AutoCloseable  it might be better to call as follows:

```
this.shutdown().get();
```

olitazl:
Good point! I added that!

alexmoore:
The RiakCluster / RiakClient objects are supposed to be longer-lived and manage a pool of connections that can be shared across multiple threads.  Are you envisioning using this like JDBC connections?

jplock:
`Closeable` extends from `Autocloseable` so I think we could just implement that - http://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html

olitazl:
@alexmoore Not only but in some cases it makes it easier to implement small projects

@jplock could be a solution

srgg:
I vote for AutoCloseable as it is more generic and modern interface, with the more generic contract.

But there is another issue, at the moment close() method is implemented as

```
public void close() throws ExecutionException, InterruptedException 
```
, whereas:


> Implementers of this interface are also strongly advised to not have the close method throw InterruptedException. This exception interacts with a thread's interrupted status, and runtime misbehavior is likely to occur if an InterruptedException is suppressed. More generally, if it would cause problems for an exception to be suppressed, the AutoCloseable.close method should not throw it.

from the [javadoc](http://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html)

olitazl:
@srgg hmm! Just catch it, doesn't seam enough! What kind of handling would you suggest?

PR #684, opened 2016-11-03
Simplification streamable commands

srgg:
Please ensure the following is present in your PR:

- [x] Unit tests for your change
- [x] Integration tests (if applicable)


srgg:
@a1 here is an API breaking change introduced by this commit https://github.com/basho/riak-java-client/pull/684/commits/b3f175bcf1ed66b4d5d6042856d3da67756bf944

Respose.Entity.getRiakObjectLocation() was renamed to getLocation().

After several next commits it turned out that Location is created on demand for each entity returned by whatever 2i query we use. IMO better to avoid such object creation especially in case when we have no understanding whether these objects useful or not. Therefore I propose to get rid of location and introduce 2 fields(Namespace, ObjectKey) instead.

src/main/java/com/basho/riak/client/api/GenericRiakCommand.java
```diff
@@ -0,0 +1,87 @@
...
+                    protected I convertQueryInfo(CoreI coreQueryInfo)
+                    {
+                        return infoConverter.convert(coreQueryInfo);
+                    }
+                };
+        coreFuture.addListener(future);
+        return future;
+    }
+
+    @SuppressWarnings("unchecked")
```
alexmoore: Most of these changes look good (and needed). 
The only thing that stands out are these two `unchecked` suppressions.  We're ignoring the type system here, which always feels bad for some reason.  Could we make these abstract methods, and then provide subclasses where `I == CoreI` and/or `R == CoreR`, and use those where necessary?  

Sort of like what I did with the `ImmediateCoreFutureAdapter`: https://github.com/basho/riak-java-client/pull/677/files#diff-1c6f3b6c65d8e0eeaee247ff938a349bR70 ?

srgg: I hate usage of  whatever suppression, therefore  it took some time from me to make a decision. The idea was to avoid  the necessity of creating additional object such as ImmediateCoreFutureAdapter.SameQueryInfo and use  Method References whenever it is possible. And also to reduce complexity I've added these default implementation that covers default scenario that doesn't require response/info conversion.

But later I realized that it might have sense  to introduce another class to handle this specific case and GenericRiakComand should extend it.

alexmoore:
+1

PR #691, opened 2016-12-07
Add TS 1.5 Features

alexmoore:
Adds the Blob data type and support for the ever-expanding DESCRIBE table result.

## Description
Adds support for the new BLOB TS data type, and accepts another column for the DESCRIBE table result. 

## Related Issue
https://bashoeng.atlassian.net/browse/CLIENTS-935

## Motivation and Context
To implement all the things!

## How Has This Been Tested?
Tests run fine against TS 1.5 RC2.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [X] Javadoc comments for any new public classes, interfaces, etc.
- [ ] A basho_docs PR for new or changed features (basho/basho_docs#???)

### DIFF
--- a/src/main/java/com/basho/riak/client/core/codec/TermToBinaryCodec.java
+++ b/src/main/java/com/basho/riak/client/core/codec/TermToBinaryCodec.java
@@ -59,7 +59,7 @@ public class TermToBinaryCodec
         return decodeTsResponse(response);
     }
 
-    public static OtpOutputStream encodeTsQueryRequest(String queryText, byte[] coverageContext)
+    public static OtpOutputStream encodeTsQueryRequest(String queryText, byte[] coverageContext, boolean queryBufferReuse)
     {
         final OtpOutputStream os = new OtpOutputStream();
         os.write(OtpExternal.versionTag); // NB: this is the reqired 0x83 (131) value
@@ -89,7 +89,7 @@ public class TermToBinaryCodec
             os.write_binary(coverageContext);
         }
 
-        os.write_boolean(false);
+        os.write_boolean(queryBufferReuse);
 
         return os;
     }
--- a/src/main/java/com/basho/riak/client/core/codec/TermToBinaryCodec.java
+++ b/src/main/java/com/basho/riak/client/core/codec/TermToBinaryCodec.java
@@ -59,7 +59,7 @@ public class TermToBinaryCodec
         return decodeTsResponse(response);
     }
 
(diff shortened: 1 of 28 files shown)


src/main/java/com/basho/riak/client/core/codec/TermToBinaryCodec.java
```diff
@@ -322,30 +324,37 @@ private static Cell parseCell(List<RiakTsPB.TsColumnDescription> columnDescripti
     {
         if (cell instanceof OtpErlangBinary)
         {
-            OtpErlangBinary v = (OtpErlangBinary) cell;
-            String s = new String(v.binaryValue(), StandardCharsets.UTF_8);
-            return new Cell(s);
+            final OtpErlangBinary v = (OtpErlangBinary) cell;
+            if (columnDescriptions.get(j).getType() == RiakTsPB.TsColumnType.VARCHAR)
```
srgg: IMO better to throw exception in case when something is incompatible, rather then silently skip it:

```
switch(columnDescriptions.get(j).getType()){
  case RiakTsPB.TsColumnType.VARCHAR://
  case RiakTsPB.TsColumnType.BLOB: //
  default:
    throw new IllegalStateException(<proper explanation>)
}

File: src/main/java/com/basho/riak/client/core/operations/ts/CreateTableOperation.java [hunk 2]
```diff
@@ -205,6 +203,12 @@ private static StringBuilder generateKeys(TableDefinition tableDefinition, int q
             {
                 sb.append(", ")
                         .append(lk.getName());
+
+                if (lk.hasKeyOrder())
+                {
+                    sb.append(" ");
```
srgg, line 209: Might be (but not mandatory) re-written like:

```
sb.append(" ")
    .append(lk.getKeyOrder().toString());
```

src/main/java/com/basho/riak/client/core/codec/TermToBinaryCodec.java
alexmoore: We do fall through to the "default" case on line 367 / 368 where we throw an exception, but I guess it could be more specific.

